PR #1, opened 2015-02-11
ctf-ir: disallow 0-bit integer, allow 1-bit signed integer

eepp:
Enforce a positive size on integer type creation, then signed or unsigned is possible for any size > 0.


jgalar:
Merged, thanks!

PR #3, opened 2015-02-20
Fix out-of-tree build for Python bindings + doc

eepp:
Signed-off-by: Philippe Proulx <eeppeliteloop@gmail.com>

### DIFF
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -1,31 +1,34 @@
-all-local: __init__.py
+NATIVEBT_PY = nativebt.py
+NATIVEBT_WRAP_C = nativebt_wrap.c
+INIT_PY = __init__.py
+INIT_PY_IN = $(srcdir)/$(INIT_PY).in
+NATIVEBT_I = $(srcdir)/nativebt.i
 
-__init__.py: __init__.py.in
+all-local: $(INIT_PY)
+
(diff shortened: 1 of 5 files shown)


simark:
I get this when I try it:
http://paste.ubuntu.com/10328444/

Is it possible that it works for you because you already have the module installed in your python path?

eepp:
No, it's not in my Python path.

I had some people testing this patch which are using Debian-based distributions, like you probably are. The main issue was that `sphinx-build` is a script in `/usr/bin`, and its interpreter is set to `/usr/bin/python`, instead of `/usr/bin/env python`, making it use Python 2 when it's the default Python interpreter (still the case on Debian/Ubuntu, I believe) instead of whatever is set as the current Python interpreter (will not work with a virtual environment either, for example).

I just updated this branch. Now, a custom Python script (`doc/bindings/python/sphinx-build.py`) invokes Sphinx's main function, but this script starts with `#!/usr/bin/env`, so that we can call

    /usr/bin/python3 sphinx-build.py

and the Python 3 Sphinx package will be used, if available (which is also verified at the `configure` stage using a new M4 macro).

Your problem, however, was unrelated to this. So please try again with a fresh source directory and report any problem. This version was tested and working on Ubuntu 14.04.

jgalar:
Merged, thanks!

PR #4, opened 2015-03-12
Add basic object system to Babeltrace

eepp:
This patch adds a basic, yet fairly complete, object system to
Babeltrace, or more specifically, to libbabeltrace. The object system,
prefixed with `bt_object`, wraps GLib, but does not expose any
GLib object, making it suitable for public APIs. The object system
features reference counting, and it has deep-copy as well as
deep-compare functions.

The provided test reaches 92.36% of `object.c`'s lines, according
to Gcov, and after having checked the report line by line, the only
blocks that are not covered are those where GLib memory allocation
fails. Valgrind's memory leak detector also shows no lost memory
during the execution of this test.

`object.h` is Doxygen-ready, with a complete file-level documentation
introducing the API, how to use it, and the best practices.

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -5,7 +5,8 @@ babeltraceinclude_HEADERS = \
 	babeltrace/iterator.h \
 	babeltrace/trace-handle.h \
 	babeltrace/list.h \
-	babeltrace/clock-types.h
+	babeltrace/clock-types.h \
+	babeltrace/objects.h
 
 babeltracectfinclude_HEADERS = \
 	babeltrace/ctf/events.h \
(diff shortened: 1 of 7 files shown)


jgalar:
Would you mind splitting this into two commits? (Feature and tests)

eepp:
> Would you mind splitting this into two commits? (Feature and tests)

Done.

jgalar:
> Done.

K.

eepp:
> K.

:sheep:

jgalar:
Merged with minor whitespace touch-ups. Thanks!

PR #5, opened 2015-03-16
ir: add bt_ctf_trace_get_clock_by_name() and tests

eepp:
Here's a sequential search for retrieving a registered clock by name in CTF IR.

I made two separate commits like you asked last time, but if this patch is small enough to squash them, let me know.


jgalar:
Merged with whitespace fixes. Thanks!

PR #6, opened 2015-03-16
ir: add bt_ctf_stream_class_get_event_class_by_id() and tests

eepp:
This is a copy of the function above, `bt_ctf_stream_class_get_event_class_by_name()`, to get an event class by ID.


jgalar:
Merged, thanks!

PR #8, opened 2015-03-19
IR: use bt_object for environment, event attributes; some fixes, and add tests for all that

eepp:
The following patches add a `bt_ctf_attributes_*()` internal API for IR. This uses the new object system (`bt_object`) to keep an array to pairs, each pair containing an attribute name and an attribute value (any object).

This internal API is reused by:

  * environment fields (currently accepts only integer and string values to match CTF v1.8)
  * event attributes (currently accepts only a specific set of attributes to match CTF v1.8)

4fa2d7c also fixes the fact that two event classes could be added to a given stream class with the same name and/or the same ID, which is illegal as per the CTF v1.8.

Tests are added for all this. Valgrind still reports 0 lost bytes for `test_ctf_writer`.

### DIFF
--- a/formats/ctf/ir/event.c
+++ b/formats/ctf/ir/event.c
@@ -47,54 +47,122 @@ int set_integer_field_value(struct bt_ctf_field *field, uint64_t value);
 
 struct bt_ctf_event_class *bt_ctf_event_class_create(const char *name)
 {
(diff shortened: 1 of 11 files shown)


jgalar:
Merged with minor changes, thanks!

PR #11, opened 2015-04-15
IR: add string field utils

eepp:
`bt_ctf_field_string_append()` and `bt_ctf_field_string_append_len()` will be useful in the CTF packet reader, when stopping the decoding process in the middle of a string.


jgalar:
Merged with a minor change in f86bada in which I replaced g_string_insert_len() by g_string_append_len(), thanks!

eepp:
> replaced g_string_insert_len() by g_string_append_len()

Good one.

PR #13, opened 2015-04-16
Fix: return event timestamp in _timestamp_at_pos

abusque:
The method, called from properties timestamp_begin and timestamp_end, always returned None (implicitly) previously.


jgalar:
Merged, thanks!

PR #15, opened 2015-05-25
add bt_ctf_field_variant_get_current_field() and tests

eepp:
`bt_ctf_field_variant_get_current_field()` returns the current selected field of a variant field. With this function, the user does not need to find the actual enumeration field used as the variant's tag and pass it to `bt_ctf_field_variant_get_field()` to retrieve the current selected field.

Another option would be to pass `NULL` to the tag parameter of `bt_ctf_field_variant_get_field()`, but I find this new function name more explicit than juggling with `NULL` params.

### DIFF
--- a/formats/ctf/ir/event-fields.c
+++ b/formats/ctf/ir/event-fields.c
@@ -652,6 +652,31 @@ end:
 	return new_field;
 }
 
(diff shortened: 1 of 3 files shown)


jgalar:
Merged, thanks!

PR #16, opened 2015-07-14
Fix: do not init. BO of enums/arrays/sequences

eepp:
See commit message of 9fde94d.

Note: other compound types could call `bt_ctf_field_type_init()` with `FALSE` also; this has no effect since they are always empty at creation time. I let them as is to keep the previous behavior.


jgalar:
Merged in master, thanks!

PR #17, opened 2015-07-14
Fix CTF IR alignment stuff

eepp:


jgalar:
Merged in master, thanks!

PR #18, opened 2015-07-16
CTF IR: consolidate reference counting functions

eepp:
See commit msg of 79b2450.

Valgrind reports 0 leaks when running `test_ctf_writer` after having merged this PR :ok_hand:.

### DIFF
--- a/formats/ctf/ir/clock.c
+++ b/formats/ctf/ir/clock.c
@@ -28,12 +28,14 @@
 
 #include <babeltrace/ctf-ir/clock-internal.h>
 #include <babeltrace/ctf-ir/utils.h>
+#include <babeltrace/ctf-ir/common-internal.h>
+#include <babeltrace/ctf-ir/ref.h>
 #include <babeltrace/ctf-writer/writer-internal.h>
 #include <babeltrace/compiler.h>
 #include <inttypes.h>
 
 static
-void bt_ctf_clock_destroy(struct bt_ctf_ref *ref);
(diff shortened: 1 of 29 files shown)


jgalar:
Good idea, but let's take the opportunity to define a base "class" for all BT objects (basically, IR and bt_object) along with documentation stating that this class _must_ be the first member of subclasses.

eepp:
Updated.

jgalar:
The latest patch set doesn't pass tests
```
ok 1 - BT_OBJECT_PUT() resets the variable to NULL
ok 2 - BT_OBJECT_MOVE() resets the source variable to NULL
ok 3 - BT_OBJECT_MOVE() moves the ownership
ok 4 - bt_object_freeze() fails with an object set to NULL
ok 5 - bt_object_freeze() succeeds with a null object
ok 6 - NULL is not frozen
ok 7 - the null singleton is frozen
ok 8 - bt_object_is_frozen() returns false with a fresh object
ok 9 - bt_object_freeze() passes with a frozen object
ok 10 - bt_object_is_frozen() returns true with a frozen object
ok 11 - bt_object_null is not NULL
ok 12 - bt_object_null is a null object
ok 13 - getting bt_object_null does not cause a crash
lt-test_bt_objects: ../include/babeltrace/ref-internal.h:61: bt_ref_put: Assertion `ref->release_func' failed.
```

jgalar:
Merged in master, thanks!

PR #22, opened 2015-07-27
IR: add type ID helpers

eepp:
Those are things we do all the time so why not providing them in the API?


jgalar:
Merged, thanks!

PR #24, opened 2015-07-30
CTF IR: make sure dynamic scopes always exist

eepp:
Dynamic scopes of event and stream classes that were previously not initialized are now initialized as empty structures (which are now valid).

This should simplify things for the future.


jgalar:
Leaving out 4ff8e552 since this could lead to ambiguities in the IR. However 3d30cc4 has been merged in master. Thanks!

PR #30, opened 2015-10-05
[RFC] Implement debug info analysis

abusque:
This changeset adds the debug information analysis to babeltrace. This analysis allows mapping between instruction pointers (IPs) from a trace event's context to function name, from either DWARF info or ELF symbols, as well as source location, i.e. source file and line number, only if DWARF info is present.

In order for the analysis to work properly, the trace has to be taken with a 2.8.0-pre (that is, the current master) of LTTng UST (for the required statedump and (optionally) libdl events), with ip and vpid contexts enabled.

There is a dependency on elfutils, for libelf and libdw.


abusque:
Before officially merging this into master, I plan to add configure script checks for the newly introduced dependencies.

abusque:
Just wanted to make it clear that the configure checks I mentioned in my previous comment have been implemented and merged into this PR.

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+{
+	int ret;
+	Dwarf_Unsigned _cu_offset;
+
+	if (!dwarf_info || !cu_offset) {
+		goto error;
+	}
+
+	ret = dwarf_next_cu_header(*dwarf_info, NULL, NULL, NULL, NULL,
+				&_cu_offset, NULL);
```
jgalar: Indent to 2 tabs

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+
+	if (!cu) {
+		goto error;
+	}
+
+	dwarf_die = g_new0(Dwarf_Die, 1);
+	if (!dwarf_die) {
+		goto error;
+	}
+	/* dwarf_siblingof on the cu fetches its root DIE. */
```
jgalar: cu -> compilation unit
DIE -> Debugging Information Entry (DIE)

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+	int ret;
+	Dwarf_Unsigned cu_offset;
+
+	if (!dwarf_info) {
+		goto error;
+	}
+
+	do {
+		ret = dwarf_next_cu_header(*dwarf_info, NULL, NULL, NULL, NULL,
+					&cu_offset, NULL);
```
jgalar: Indent to 2 tabs

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+	cu = g_new0(struct durin_cu, 1);
+	if (!cu) {
+		goto error;
+	}
+	cu->dwarf_info = dwarf_info;
+	cu->offset = 0;
+
+	return cu;
+
+error:
```
jgalar: No need for this label if we set `cu = NULL`.

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+	/* dwarf_siblingof on the cu fetches its root DIE. */
+	ret = dwarf_siblingof(*cu->dwarf_info, NULL, dwarf_die, &error);
+	if (ret != DW_DLV_OK) {
+		goto error;
+	}
+	die = g_new0(struct durin_die, 1);
+	if (!die) {
+		goto error;
+	}
+	die->cu = cu;
```
jgalar: Indicate that `durin_die` takes no ownership of `cu`.

formats/ctf-text/durin.c
compudj: We could return NULL in every error branch too, but having a single exit point is a nice property for a function. So here the idea is to have one exit point for OK and one exit point for all error.

jgalar: I meant setting `cu = NULL` at the declaration. Maybe we're saying the same thing?
I would not tend to have an error path if there is no need for one (e.g. to perform some clean-up).

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+
+void durin_die_destroy(struct durin_die *die)
+{
+	if (!die) {
+		return;
+	}
+
+	if (die->dwarf_die) {
+		dwarf_dealloc(*die->cu->dwarf_info, *die->dwarf_die,
+			      DW_DLA_DIE);
```
jgalar: tabs/space mix-up in indentation.

formats/ctf-text/durin.c
compudj: I'm OK with either returning NULL everywhere or goto error path to return NULL in this case. It's just a matter of where we came from in the review process: earlier versions of this work did forget about doing clean-up in error paths, so I told Antoine to stick to a 2-return paths for _everything_, even if there is nothing to cleanup in the error path.

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+		g_free(die->dwarf_die);
+	}
+	g_free(die);
+}
+
+int durin_die_next(struct durin_die *die)
+{
+	int ret;
+	Dwarf_Die *next_die = NULL;
+	Dwarf_Error error;
```
jgalar: I'm surprised this didn't clash with the "error" label since GCC allows the use of lables as values...

https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+		goto error;
+	}
+
+	next_die = g_new0(Dwarf_Die, 1);
+	if (!next_die) {
+		goto error;
+	}
+
+	if (die->depth == 0) {
+		ret = dwarf_child(*die->dwarf_die, next_die, &error);
```
jgalar: It is weird that the same `die` is changed to point to another `die` vs. implementing an iterator, which would seem more appropriate here.

Why is depth limited to 1? (as opposed to traversing the entire tree)

formats/ctf-text/durin.c
abusque: I wasn't aware of that at. I haven't found any info in the doc explaining how it could avoid the clash, but seems like it does.

formats/ctf-text/durin.c
```diff
@@ -0,0 +1,334 @@
...
+		goto error;
+	}
+
+	return 0;
+
+error:
+	return -1;
+}
+
+struct durin_cu *durin_cu_create(Dwarf_Debug *dwarf_info)
```
jgalar: Even though it has no side-effect, we typically indicate a function's visibility in the implementation file (i.e. BT_HIDDEN).

formats/ctf-text/durin.c
compudj: Those are probably two different namespaces.

Taking a value from a label requires the operator &&, and goto *value requires the "*" operator.

formats/ctf-text/durin.c
abusque: Concerning the limited depth, see the comment at `include/babeltrace/ctf-text/durin.h:60`. What should be made explicit in that the so-called "interesting" DIEs are those whose tag value is `DW_TAG_subprogram`.

abusque: As for the iterator, I think this alternative approach was simply to mirror the equivalent functions from libdwarf/libdw, but there is no real technical reason to not have an iterator, as far as I can recall.

formats/ctf-text/durin.c
abusque: I was under the impression that it was to be done in the header files. Unless you mean putting it in both places?

jgalar: Yes, both.
Like I said, it's only used as documentation in the `.c` file.

formats/ctf-text/durin.c
jgalar: Okay, thanks. Not a deal breaker since this is internal.

include/babeltrace/ctf-text/durin.h
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Get a DIE's name.
+ *
+ * @param die	durin_die instance
+ * @param name	Out parameter, the DIE's name
+ * @returns	0 on succes, -1 on failure
+ */
+BT_HIDDEN
+int durin_die_get_name(struct durin_die *die, char **name);
```
jgalar: Add comment that the caller does not need to free `name`.
Looking at libdwarf's source code seems to indicate that the string is valid as long as the underlying `dwarf_die` is not `free`'d. This also means that `durin_die_next()` must not be called.

formats/ctf-text/so-info.c
```diff
@@ -0,0 +1,756 @@
...
+
+/*
+ * An adress printed in hex is at most 20 bytes (16 for 64-bits +
+ * leading 0x + optional leading '+' if addr is an offset + null
+ * character).
+ */
+#define ADDR_STR_LEN 20
+
+struct so_info *so_info_create(const char *path, uint64_t low_addr,
+			uint64_t memsz)
```
jgalar: Indent to 2 tabs.

formats/ctf-text/so-info.c
```diff
@@ -0,0 +1,756 @@
...
+	if (!path) {
+		goto error;
+	}
+
+	so = g_new0(struct so_info, 1);
+	if (!so) {
+		goto error;
+	}
+
+	so->path = path;
```
jgalar: Copy path to decouple `so`'s lifetime from the `path` argument.

formats/ctf-text/so-info.c
```diff
@@ -0,0 +1,756 @@
...
+	if (!so) {
+		return;
+	}
+
+	if (so->dwarf_info) {
+		dwarf_finish(*so->dwarf_info, NULL);
+		g_free(so->dwarf_info);
+	}
+	g_free(so->build_id);
+	g_free(so->dbg_link_filename);
```
jgalar: `g_free(path);`

converter/babeltrace.c
```diff
@@ -175,6 +189,10 @@ static void usage(FILE *fp)
 	fprintf(fp, "      --clock-gmt                Print clock in GMT time zone (default: local time zone)\n");
 	fprintf(fp, "      --clock-force-correlate    Assume that clocks are inherently correlated\n");
 	fprintf(fp, "                                 across traces.\n");
+#ifdef ENABLE_DEBUGINFO
+	fprintf(fp, "      --debug-dir                Directory in which to look for debugging information\n");
+	fprintf(fp, "                                 files during debuginfo analysis. (default: /usr/lib/debug/)\n");
```
jgalar: I think the default should be $prefix/lib/debug

abusque: I get your point and I would usually agree, but this doesn't seem to be the default behaviour adopted by GDB for its own lookup. See https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html

abusque: So I guess it's just a question of whether we want to be consistent with how GDB behaves, or provide a default that accommodates users with different `$prefix` values.

abusque: Also, tangentially related, here's some info on how Fedora handles debuginfo through its packaging system: https://fedoraproject.org/wiki/Packaging:Debuginfo

jgalar: Urgh... :(
Makes sense then

converter/babeltrace.c
```diff
@@ -70,6 +75,9 @@ static char *opt_input_format, *opt_output_format;
  */
 static GPtrArray *opt_input_paths;
 static char *opt_output_path;
+#ifdef ENABLE_DEBUGINFO
```
jgalar: I would only put the "ifdef" directives where they are absolutely necessary. In this case, that would be the `#include <babeltrace/debuginfo.h>` and the entry in the `long_options` array. The rest will not really have an impact (`free(NULL)`) and should keep the code more readable.

converter/babeltrace.c
```diff
@@ -683,6 +709,12 @@ int main(int argc, char **argv)
 	if (opt_output_format)
 		strlower(opt_output_format);
 
+#ifdef ENABLE_DEBUGINFO
+	if (opt_debug_dir) {
+		debug_info_set_debug_dir(opt_debug_dir);
```
jgalar: `debug_info_set_debug_dir()` should be stubbed to have no effect if `ENABLE_DEBUGINFO` is not defined.

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
+Babeltrace Debug Info Analysis
+-----------------------------
+
+The babeltrace debug info analysis is a set of new features which
```
jgalar: Remove "new" (they will not always be new).

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+allow mapping events from a trace to their location in source code.
```
jgalar: mapping the "ip" field of events to their location in source code.

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+Prerequisites
+-------------
+
+In order to install a version of babeltrace with debug info support,
+the following software is required:
```
jgalar: software is -> libraries are

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+    * libelf
+    * libdw
+
+The two libraries required are provided by the elfutils project
+(https://fedorahosted.org/elfutils/), and can be installed through the
+elfutils package on Ubuntu, Debian, RHEL, and others.
+
+Getting Babeltrace with Debug Info Support
```
jgalar: Remove this section since this will be mainline :)

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+be found in the README file.
+
+Compiling for Debug Info Analysis
+---------------------------------
+
+Traced programs on which debug info analysis is to be performed can be
+compiled in a few different ways, and still lead to useful results.
+
+Ideally, one should compile the program in debug mode, using simply
+the `-g` flag. This generates DWARF information, which is then used by
```
jgalar: "produces debugging information in the operating system's native format".
We should mention we only support DWARF 2

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+name, however, the analysis will give as a result the name of the
+nearest function symbol, plus an offset in bytes to the location in
+the executable from which the event originated.
+
+If the executable has neither ELF symbols nor DWARF information,
+babeltrace will be unable to map an event to its source location, and
+will simply display the instruction pointer (address), as in prior
+versions of babeltrace.
+
+Getting the Right Tracer
```
jgalar: fix capitalization

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+Ideally, one should compile the program in debug mode, using simply
+the `-g` flag. This generates DWARF information, which is then used by
+babeltrace to map an event's source location to a file and line
+number, and the name of the surrounding function.
+
+If the executable is not compiled with `-g` enabled, and thus no DWARF
+information is available, babeltrace will use ELF symbols from the
+executable. Instead of providing source file, line number and function
+name, however, the analysis will give as a result the name of the
```
jgalar: give as a result -> provide

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+Debug info analysis is performed automatically by babeltrace, provided
+the trace contains sufficient information. In order to be able to
+trace all the necessary information, the following software is
+required:
+
+    * lttng-ust version 2.8.0-pre or later
+    * lttng-tools, corresponding version
+
+LTTng 2.8 is currently in development, which is why the pre-release
```
jgalar: Now released.

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+    * https://launchpad.net/~lttng/+archive/ubuntu/daily
+
+Tracing for Debug Info Analysis
+-------------------------------
+
+Babeltrace needs some extra information from contexts, namely ip and
+vpid, to perform its analysis. These can be enabled after the creation
+of a tracing session as follows:
+
+   $ lttng add-context -u -t ip -t vpid
```
jgalar: Prefer using long-form options to increase readability.

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+The tracing can now be performed as it normally would. Once the trace
```
jgalar: now -> then

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+Analysing the Trace
+-------------------
+
+To perform the analysis, the trace can simply be read as it normally
+would:
+
+   $ babeltrace <path/to/trace>
+
+Debug info analysis is on by default, and will automatically print the
```
jgalar: remove comma

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+extra source location information if it can find it. A sample output
```
jgalar: "find it" -> "be provided"

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+may look like this:
+
+    [...]
+    [19:53:41.648394410] (+0.000022616) colossus my_provider:my_first_tracepoint: { cpu_id = 2 }, { ip = 0x7F75AEDFB4D4, debug_info = { func = "foo", source_loc = "/home/efficios/example/libhello.c:7" }, vpid = 25577 }, { my_string_field = "hello, tracer", my_integer_field = 42 }
```
jgalar: Update if we decide to change the output format to match Trace Compass.

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+    [19:53:41.648423786] (+0.000029376) colossus my_provider:my_first_tracepoint: { cpu_id = 2 }, { ip = 0x7F75AEDFB5A6, debug_info = { func = "bar", source_loc = "/home/efficios/example/libhello.c:13" }, vpid = 25577 }, { my_string_field = "recoltes et semailles", my_integer_field = 57 }
+    [19:53:41.648435726] (+0.000011940) colossus my_provider:my_other_tracepoint: { cpu_id = 2 }, { ip = 0x7F75AEDFB66B, debug_info = { func = "baz", source_loc = "/home/efficios/example/libhello.c:20" }, vpid = 25577 }, { some_field = 1729 }
+    [...]
+
+The interesting part is the debug_info section of the context:
+
+    debug_info = { func = "foo", source_loc = "/home/efficios/example/libhello.c:7" }
```
jgalar: I think `source_loc` should be compact by default
`source_loc = "libhello.c:7"`

@eepp: I would add a "long path" option, but in BT 1.4, that would mean introducing another "global" option... We can leave it as-is in 1.4 and introduce a "ctf-text" option in 2.x. What do you think?

eepp: I don't care about global boolean options that apply to a specific plugin, it's easy to handle in BT 2.0's config "translator". If it's in 1.4, I would allow it in 2.0 too.

doc/debuginfo.txt
abusque: I know this is pretty trivial but I just thought I'd justify regardless: I usually follow title case rules as specified by both the MLA and APA style guides, which usually require capitalizing all non-article words in headings.

eepp: > fix capitalization

What's the problem here?

jgalar: @abusque Alright then.
@eepp None, as indicated by @abusque.

doc/debuginfo.txt
jgalar: I have no doubt it's easy to handle, but I'm looking at it from a uniformity/UX standpoint.

I'm hesitating between either

1) Showing full paths in 1.4 and switching to the compact form in 2.x + introduce an output option specific to the ctf-text plug-in.

2) Adding a global option right away. Not sure about the name though. --full-path (would apply to all paths shown by BT?), --debug-info-full-path?

Pensées?

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+
+The debug information now provides only the function name as given by
+ELF symbols, plus an offset in bytes from the symbol to the actual
+location of the tracepoint instance.
+
+Debug Info and Dynamic Loading
+------------------------------
+
+Babeltrace can resolve addresses of events originating from
+dynamically loaded libraries, provided some extra information is
```
jgalar: remove comma

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+    $ LD_PRELOAD="liblttng-ust-dl.so" <path/to/executable>
+
+The tracing and analysis can now be performed as described in prior
+sections, and events from tracepoints in dlopened libraries will be
+resolved automatically by babeltrace.
+
+Separate Debug Info
+-------------------
+
+It is possible to have DWARF debug information separate from an
```
jgalar: store DWARF debug information separately from

doc/debuginfo.txt
eepp: The default behaviour (no parameters) should be the same in 1.4 and 2.0.

I like option 2, and `--debug-info-full-path` seems reasonable. Keep in mind that it's probably the very last new command-line option to Babeltrace.

jgalar: Yeah, I prefer option 2 too. Sounds good. :palm_tree:

doc/debuginfo.txt
```diff
@@ -0,0 +1,187 @@
...
+--------------------
+
+When performing debug info analysis, babeltrace uses an executable's
+path from when it was traced to resolve debug information. If,
+however, the trace was taken on a different machine, or the executable
+was simply moved since the trace was taken, it is possible to override
+the base path from which to look for the debug information. To do so,
+the --debug-dir babeltrace command-line flag is available.
+
+If a --debug-dir value is given, it will replace the default
```
jgalar: @eepp Any thoughts on this parameter then? `--debug-info-dir`?

formats/ctf/ctf.c
```diff
@@ -2413,8 +2414,15 @@ struct bt_trace_descriptor *ctf_open_trace(const char *path, int flags,
 		goto error;
 	}
 
+	ret = trace_debug_info_create(td);
+
```
jgalar: Remove new line.

formats/ctf/ctf.c
```diff
@@ -2583,6 +2591,12 @@ struct bt_trace_descriptor *ctf_open_mmap_trace(
 	if (ret)
 		goto error_free;
 
+	ret = trace_debug_info_create(td);
+
```
jgalar: Remove new line.

include/babeltrace/debuginfo-wrapper.h
```diff
@@ -0,0 +1,130 @@
...
+#include <babeltrace/ctf-ir/metadata.h>
+
+#ifdef ENABLE_DEBUGINFO
+
+#include <babeltrace/debuginfo.h>
+#include <babeltrace/ctf-text/types.h>
+
+static inline
+void ctf_text_integer_write_debug_info(struct bt_stream_pos *ppos,
+				       struct bt_definition *definition)
```
jgalar: Function prototype indentations need to be adjusted in this file.

include/babeltrace/crc32.h
```diff
@@ -0,0 +1,128 @@
...
+	0xbdbdf21c, 0xcabac28a, 0x53b39330, 0x24b4a3a6,
+	0xbad03605, 0xcdd70693, 0x54de5729, 0x23d967bf,
+	0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94,
+	0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d,
+};
+
+/**
+ * Compute a 32-bit cyclic redundancy checksum for a given file.
+ *
+ * On sucess, the out parameter crc is set with the computed checksum
```
jgalar: sucess -> success

include/babeltrace/crc32.h
```diff
@@ -0,0 +1,128 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <babeltrace/babeltrace-internal.h>
+
+#define CRC(crc, ch)	 (crc = (crc >> 8) ^ crctab[(crc ^ (ch)) & 0xff])
```
jgalar: This macro and the constants below only seem to be used by the `crc32()` function. Move them to crc32.c.

include/babeltrace/debuginfo-wrapper.h
```diff
@@ -0,0 +1,130 @@
...
+	if (strcmp(trace->env.domain, "ust") != 0) {
+		goto end;
+	}
+
+	if (strcmp(trace->env.tracer_name, "lttng-ust") != 0) {
+		goto end;
+	}
+
+	trace->debug_info = debug_info_create();
+
```
jgalar: Remove new line.

include/babeltrace/debuginfo-wrapper.h
```diff
@@ -0,0 +1,130 @@
...
+	debug_info_destroy(trace->debug_info);
+}
+
+static inline
+void ctf_init_debug_info()
+{
+	int ret;
+
+	ret = debug_info_init();
+	assert(!ret);
```
jgalar: Why not propagate the error to the caller?

include/babeltrace/debuginfo.h
```diff
@@ -0,0 +1,29 @@
...
+#define _BABELTRACE_DEBUGINFO_H
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+struct debug_info_source {
+	/* Strings are owned by this */
+	char *func;
+	long unsigned int line_no;
```
jgalar: Prefer using fixed-size types (e.g. uint64_t) when overflow may occur on word-size types.
However, a 32-bit unsigned integer should be enough for files shorter than 4,294,967,295 lines. :-)

include/babeltrace/durin.h
```diff
@@ -0,0 +1,218 @@
...
+#include <stdlib.h>
+#include <dwarf.h>
+#include <elfutils/libdw.h>
+#include <babeltrace/babeltrace-internal.h>
+
+/*
+ * Durin is a wrapper over libdw providing a nicer, higher-level
+ * interface, to access basic debug information. Durin is also the
+ * eldest of the seven Fathers of the Dwarves created by the Vala Aulë
+ * in the lore of Tolkien's Middle-earth.
```
jgalar: Clever ;-)

abusque: I have no qualms against renaming it `bt_dwarf` or something equivalent, but I needed a development name.

The equivalent ELF code eventually merged into UST had an equally nerdy Middle-Earth name of Elwë (well, elwe in the source). You can probably guess its meaning.

:neckbeard:

jgalar: `bt_dwarf` is indeed more aligned with the rest of the codebase :)

include/babeltrace/debuginfo-wrapper.h
abusque: No reason, I just moved this here from the `ctf.c` constructor where it initially asserted, I'll just return the value here and `return 0` in the default/disabled implementation instead, and assert there.

include/babeltrace/durin.h
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Advance the debug information entry `die` to its first child, if
+ * any.
+ *
+ * @param die	durin_die instance
+ * @returns	0 on success, -1 on failure
+ */
+BT_HIDDEN
+int durin_die_child(struct durin_die *die);
```
jgalar: This seems superfluous since we only allow access to the first level in `durin_die_next()`.

include/babeltrace/debuginfo.h
abusque: I'll have to double check but the 64-bit requirement was from the DWARF spec and that's what `libdw` returns. I'll make the switch to `uint64_t`, though.

include/babeltrace/durin.h
eepp: > bt_dwarf is indeed more aligned with the rest of the codebase :)

Where's the fun in that! Why not `bt_durin`? Projects often come up with custom names for subsystems, like GCC's _GIMPLE_, and IMO I think it adds personality to the codebase.

include/babeltrace/durin.h
abusque: Access to DIEs at deeper level may be required when dealing with unwinding of inlined functions, for instance, which is where this function comes into play.

include/babeltrace/durin.h
jgalar: To be fair, "personality" is not something I look for in a codebase.

I have no idea what _GIMPLE_ means, which is not a good thing. At least with `bt_dwarf` I immediately know "oh, it's that thing that deals with dwarf". (assuming you know what DWARF is... but at least DWARF is a well-known standard)

eepp: Yes but then you allow more people to understand this codebase...

formats/ctf-text/ctf-text.c
```diff
@@ -237,7 +238,7 @@ const char *print_loglevel(int value)
 
 static
 int ctf_text_write_event(struct bt_stream_pos *ppos, struct ctf_stream_definition *stream)
-			 
+
```
jgalar: Remove new line.

converter/babeltrace.c
abusque: Noted, also left the `#ifdef` in `usage()` for obvious reasons.

lib/so-info.c
```diff
@@ -0,0 +1,1223 @@
...
+ *
+ * More specifically, the parameter `die` is expected to be a
+ * subprogram (function) DIE, and this predicate tells whether any
+ * subroutines are inlined within this function and would contain
+ * `addr`.
+ *
+ * Do note that this function advances the position of `die`. If the
+ * address is found within one of its children, `die` will be pointing
+ * to that child upon returning from the function, allowing to extract
+ * info deemed necessary.
```
jgalar: info -> the info (or information)

include/babeltrace/durin.h
abusque: This function (as well as `durin_die_next` and others) should distinguish between a hard error (i.e. failed allocation) and no child DIE available by returning -1 for the former and 1 for the latter.

doc/debuginfo.txt
abusque: I feel this is a correct use of a comma before a subordinate clause that introduces a conditional, although I did go ahead and replace "provided" by "provided that".

doc/debuginfo.txt
abusque: I did change "have" to "store", but I feel that "separate" is the proper usage here as it serves the role of adjective as in "separate from an executable", rather than that of adverb, as if the emphasis were on "to store separately".

tests/lib/test_durin.c
```diff
@@ -0,0 +1,107 @@
...
+	fd = open(path, O_RDONLY);
+	ok(fd >= 0, "Open DWARF file %s", path);
+	dwarf_info = dwarf_begin(fd, DWARF_C_READ);
+	ok(dwarf_info != NULL, "dwarf_begin succesful");
+	cu = durin_cu_create(dwarf_info);
+	ok(cu != NULL, "durin_cu_create succesful");
+	ret = durin_cu_next(cu);
+	ok(ret == 0, "durin_cu_next succesful");
+	die = durin_die_create(cu);
+	ok(die != NULL, "durin_die_create succesful");
```
jgalar: succesful -> successful

tests/lib/test_durin.c
```diff
@@ -0,0 +1,107 @@
...
+	/*
+	 * Test durin_die_next twice, as the code path is different
+	 * for DIEs at depth 0 (just created) and other depths.
+	 */
+	ret = durin_die_next(die);
```
jgalar: Add an `assert(die)` to silence linter warnings.

formats/ctf/ctf.c
```diff
@@ -2583,6 +2590,11 @@ struct bt_trace_descriptor *ctf_open_mmap_trace(
 	if (ret)
 		goto error_free;
 
+	ret = trace_debug_info_create(td);
+	if (ret) {
+		goto error;
```
jgalar: This will leak `td`, `goto error_free`

jgalar:
The PR is missing modifications to the man page.

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+	uint8_t *build_id;
+	size_t build_id_len;
+	/* Optional debug link info. */
+	char *dbg_link_filename;
+	uint32_t dbg_link_crc;
+	/* FDs to ELF and DWARF files. */
+	int elf_fd;
+	int dwarf_fd;
+	/* Denotes whether the executable is position independent code. */
+	uint8_t is_pic : 1;
```
jgalar: Switched to `bool`.
Also, omit spaces before and after the colon

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+ *			-1 on failure
+ */
+static inline
+int so_info_has_address(struct so_info *so, uint64_t addr)
+{
+	if (!so) {
+		return -1;
+	}
+
+	return addr >= so->low_addr && addr <= so->high_addr;
```
jgalar: Is the high_addr really inclusive?
I see `so->high_addr = so->low_addr + so->memsz;` in so-info.c:136 which makes it seem like it is an exclusive upper-bound.

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+	int elf_fd;
+	int dwarf_fd;
+	/* Denotes whether the executable is position independent code. */
+	uint8_t is_pic : 1;
+	/* Denotes whether the SO only has ELF symbols and no DWARF info. */
+	uint8_t is_elf_only : 1;
+};
+
+struct source_location {
+	long long unsigned int line_no;
```
jgalar: Switch to uint64_t

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+
+/**
+ * Get the name of the function containing a given address within an
+ * executable.
+ *
+ * If no DWARF info is available, the function falls back to ELF
+ * symbols and the "function name" is in fact the name of the closest
+ * symbol, followed by the offset between the symbol and the address.
+ *
+ * On success, if found, the out parameter `func_name` is set. On
```
jgalar: What is the lifetime of `func_name`?

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+ * @param addr		Virtual memory address for which to find the
+ *			source location
+ * @param src_loc	Out parameter, the source location
+ * @returns		0 on success, -1 on failure
+ */
+BT_HIDDEN
+int so_info_lookup_source_location(struct so_info *so, uint64_t addr,
+		struct source_location **src_loc);
+
+
```
jgalar: Remove extra new line.

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+	}
+
+	cu = durin_cu_create(so->dwarf_info);
+	if (!cu) {
+		goto error;
+	}
+
+	while (durin_cu_next(cu) == 0) {
+		int ret;
+		ret = so_info_lookup_cu_src_loc(cu, addr, &_src_loc);
```
jgalar: Add new line between variable declarations and code.

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+		char **func_name);
+
+/**
+ * Get the source location (file name and line number) for a given
+ * address within an executable.
+ *
+ * If no DWARF info is available, the source location cannot be found
+ * and the function will return unsuccesfully.
+ *
+ * On success, if found, the out parameter `src_loc` is set. On
```
jgalar: What is the lifetime of `src_loc`?

lib/debuginfo.c
```diff
@@ -0,0 +1,734 @@
...
+	GHashTable *baddr_to_so_info;
+
+	/* Hash table: IP to (struct debug_info_source *); owned by this */
+	GHashTable *ip_to_debug_info_src;
+};
+
+struct debug_info {
+	/*
+	 * Hash table of VPIDs (int64_t) to (struct ctf_proc_debug_infos*);
+	 * owned by this.
```
jgalar: this -> `debug_info`

lib/debuginfo.c
```diff
@@ -0,0 +1,734 @@
...
+static GQuark q_statedump_debug_link;
+static GQuark q_statedump_build_id;
+static GQuark q_statedump_start;
+static GQuark q_dl_open;
+
+struct proc_debug_info_sources {
+	/* Hash table: base address to so info; owned by this */
+	GHashTable *baddr_to_so_info;
+
+	/* Hash table: IP to (struct debug_info_source *); owned by this */
```
jgalar: this -> `proc_debug_info_sources`

lib/debuginfo.c
```diff
@@ -0,0 +1,734 @@
...
+
+	debug_info_src = g_new0(struct debug_info_source, 1);
+
+	if (!debug_info_src) {
+		goto end;
+	}
+
+	/* Lookup function name */
+	ret = so_info_lookup_function_name(so, ip, &debug_info_src->func);
+
```
jgalar: Remove extra new line.

lib/debuginfo.c
```diff
@@ -0,0 +1,734 @@
...
+	}
+
+	/* Can't retrieve src_loc from ELF only, skip it */
+	if (so->is_elf_only) {
+		goto end;
+	}
+
+	/* Lookup source location */
+	ret = so_info_lookup_source_location(so, ip, &src_loc);
+
```
jgalar: Remove extra new line.

lib/debuginfo.c
```diff
@@ -0,0 +1,734 @@
...
+	struct proc_debug_info_sources *proc_dbg_info_src = NULL;
+
+	/* Exists? Return it */
+	proc_dbg_info_src = g_hash_table_lookup(ht, key);
+
+	if (proc_dbg_info_src) {
+		goto end;
+	}
+
+	/* Create it */
```
jgalar: Not sure how useful this comment is.

include/babeltrace/debuginfo.h
```diff
@@ -0,0 +1,29 @@
...
+};
+
+struct debug_info;
+struct ctf_event_definition;
+
+int debug_info_init(void);
+
+struct debug_info *debug_info_create(void);
+
+void debug_info_set_debug_dir(char *path);
```
jgalar: Mention that no ownership of `path` is taken.

include/babeltrace/so-info.h
```diff
@@ -0,0 +1,209 @@
...
+ * anything else.
+ *
+ * @returns		0 on success, -1 on failure
+ */
+BT_HIDDEN
+int so_info_init(void);
+
+/**
+ * Sets the so_info framework's debug directory, in which to look for
+ * separate debug info.
```
jgalar: Mention that no ownership of `path` is taken.

lib/durin.c
```diff
@@ -0,0 +1,352 @@
...
+	if (!dwarf_info) {
+		goto error;
+	}
+
+	cu = g_new0(struct durin_cu, 1);
+	if (!cu) {
+		goto error;
+	}
+	cu->dwarf_info = dwarf_info;
+	cu->offset = 0;
```
jgalar: `g_new0()` takes care of that.

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+	 * Only set dwarf_info the first time it is read, to avoid
+	 * reading it uselessly.
+	 */
+	so->dwarf_info = NULL;
+	so->dwarf_path = NULL;
+	so->build_id = NULL;
+	so->build_id_len = 0;
+	so->dbg_link_filename = NULL;
+	so->dbg_link_crc = 0;
+	so->is_elf_only = 0;
```
jgalar: Handled by `g_new0()`

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+	build_id_file_len = (2 * so->build_id_len) + 2;
+	build_id_file = malloc(build_id_file_len);
+	if (!build_id_file) {
+		goto error;
+	}
+
+	snprintf(build_id_file, 4, "%02x/", so->build_id[0]);
+	for (i = 1; i < so->build_id_len; ++i) {
+		int path_idx = 3 + 2 * (i - 1);
+		snprintf(&build_id_file[path_idx], 3, "%02x", so->build_id[i]);
```
jgalar: Add new line to separate variable declarations from code.

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+					die, addr);
+				goto end;
+			}
+		}
+	}
+
+end:
+	if (found) {
+		char *filename = NULL;
+		long unsigned int line_no;
```
jgalar: uint64_t

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+		struct source_location **src_loc)
+{
+	struct source_location *_src_loc = NULL;
+	struct durin_die *die = NULL;
+	const char *filename = NULL;
+	Dwarf_Line *line = NULL;
+	Dwarf_Addr line_addr;
+	int ret, line_no;
+
+
```
jgalar: Remove extra newline.

lib/so-info.c
```diff
@@ -0,0 +1,1225 @@
...
+	struct source_location *_src_loc = NULL;
+
+	if (!so || !src_loc) {
+		goto error;
+	}
+
+	/* Set DWARF info if it hasn't been accessed yet. */
+	if (!so->dwarf_info && !so->is_elf_only) {
+		if (so_info_set_dwarf_info(so)) {
+			/* Failed to set DWARF info . */
```
jgalar: Remove space between `info` and `.`

include/babeltrace/so-info.h
jgalar: I have also changed the corresponding test.

lib/Makefile.am
```diff
@@ -13,7 +13,16 @@ libbabeltrace_la_SOURCES = babeltrace.c \
 			   values.c \
 			   ref.c
 
-libbabeltrace_la_LDFLAGS = -version-info $(BABELTRACE_LIBRARY_VERSION)
+libbabeltrace_la_LDFLAGS = -version-info $(BABELTRACE_LIBRARY_VERSION) \
+	-lelf -ldw
```
jgalar: Should be conditional (as below).

abusque: Rebasing mistake, I'll fix it right now.

lib/debuginfo.c
```diff
@@ -620,6 +622,23 @@ void handle_statedump_soinfo_event(struct debug_info *debug_info,
...
+			goto end;
+		}
+
+		if (is_pic_def->declaration->id != BT_CTF_TYPE_ID_INTEGER) {
+			goto end;
+		}
+
+		is_pic = (bt_get_unsigned_int(is_pic_def) == 1);
+	} else {
+		/* dlopen has no is_pic field, because the shared
```
jgalar: Use comments of the form
```
/*
 * Hello.
 */
```

lib/so-info.c
```diff
@@ -704,16 +752,11 @@ int so_info_lookup_elf_function_name(struct so_info *so, uint64_t addr,
 			goto error;
 		}
 
-		snprintf(offset_str, ADDR_STR_LEN, "+%#0" PRIx64,
-				addr - sym->st_value);
-		_func_name = malloc(strlen(sym_name) + ADDR_STR_LEN);
-		if (!_func_name) {
+		ret = so_info_append_offset_str(sym_name, sym->st_value, addr,
+						func_name);
```
jgalar: Indentation (2 tabs).

lib/so-info.c
```diff
@@ -791,17 +832,11 @@ int so_info_lookup_cu_function_name(struct bt_dwarf_cu *cu, uint64_t addr,
 			goto error;
 		}
 
-		snprintf(offset_str, ADDR_STR_LEN, "+%#0" PRIx64,
-				addr - low_addr);
-		_func_name = malloc(strlen(die_name) + ADDR_STR_LEN);
-		if (!_func_name) {
+		ret = so_info_append_offset_str(die_name, low_addr, addr,
+						func_name);
```
jgalar: Indentation (2 tabs).

abusque: Both of these were already fixed at the time of your comment in a later commit. All of these are gonna get squashed prior to merging.

abusque: That way no intermediate bad indentation should remain.

jgalar: Alright.

jgalar:
Do you still have the files used to generate `debuginfo-data` (tests)?

abusque:
Yes, should I add them to the tree?

jgalar:
Yes, please. :smile:

jgalar:
Merged, thanks!

abusque:
Thanks! I noticed the output examples in `doc/debuginfo.txt` need updating, I'll submit a patch shortly.

PR #34, opened 2016-02-23
test_ctf_ir_ref.c: test all permutations of put

eepp:
This one adds a function in `test_ctf_ir_ref.c` which tests all the possible permutations of putting the user references of a CTF IR system made of:

  * A CTF writer
  * The trace of this CTF writer
  * A stream class added to this trace
  * A stream created from this stream class
  * An event class added to this stream class
  * An event created from this event class and appended to this stream

The patch does not add a TAP test per se, but the executable can be given to Valgrind to verify that there's no leak. In the future, if the reference counting system is changed somehow, this test could crash, which would be a good thing I guess.

WDYT?

### DIFF
--- a/tests/lib/test_ctf_ir_ref.c
+++ b/tests/lib/test_ctf_ir_ref.c
@@ -20,21 +20,40 @@
  */
 
 #include "tap/tap.h"
+#include <babeltrace/ctf-writer/writer.h>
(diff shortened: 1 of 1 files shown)


jgalar:
Merged, thanks!

PR #37, opened 2016-04-06
[RFC] Stream intersect bindings

abusque:
Initial implementation of python bindings for stream intersection mode. This adds a function to the `bt_ctf_iter` API, previously located within the converter code, for reuse between the converter and the bindings, as well as an `intersect_mode` attribute to the python class `TraceCollection`. The default behaviour is to set this attribute to `False`, which means existing python code will behave exactly the same.

### DIFF
--- a/bindings/python/nativebt.i
+++ b/bindings/python/nativebt.i
@@ -113,6 +113,11 @@ int _bt_python_ctf_clock_get_uuid_index(struct bt_ctf_clock *clock,
 int _bt_python_ctf_clock_set_uuid_index(struct bt_ctf_clock *clock,
 		size_t index, unsigned char value);
 struct bt_iter_pos *_bt_python_create_iter_pos(void);
+struct bt_ctf_iter *_bt_python_ctf_iter_create_intersect(
+	struct bt_context *ctx,
+	struct bt_iter_pos *inter_begin_pos,
+	struct bt_iter_pos *inter_end_pos);
+int _bt_python_has_intersection(struct bt_context *ctx);
 
 /* context.h, context-internal.h */
 %rename("_bt_context_create") bt_context_create(void);
--- a/bindings/python/python-complements.c
+++ b/bindings/python/python-complements.c
@@ -33,6 +33,8 @@
 #include <babeltrace/ctf-ir/event-class.h>
 #include <babeltrace/ctf-ir/clock-internal.h>
 #include <babeltrace/iterator.h>
+#include <babeltrace/ctf/iterator.h>
+#include <babeltrace/ctf/events-internal.h>
 #include <glib.h>
 
 /* List-related functions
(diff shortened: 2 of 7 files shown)


bindings/python/reader.py
```diff
@@ -219,7 +222,18 @@ def _timestamp_at_pos(self, pos_ptr):
         return ev.timestamp
 
     def _events(self, begin_pos_ptr, end_pos_ptr):
-        ctf_it_ptr = nbt._bt_ctf_iter_create(self._tc, begin_pos_ptr, end_pos_ptr)
+        if self.intersect_mode:
+            has_intersection = nbt._bt_python_has_intersection(self._tc)
+            if not has_intersection:
+                raise ValueError("No intersection found between trace files.")
```
jgalar: I'm not sure about throwing an exception here. If there is no intersection, I'd be tempted to return nothing; it may not be an error.

bindings/python/nativebt.i
```diff
@@ -113,6 +113,11 @@ int _bt_python_ctf_clock_get_uuid_index(struct bt_ctf_clock *clock,
 int _bt_python_ctf_clock_set_uuid_index(struct bt_ctf_clock *clock,
 		size_t index, unsigned char value);
 struct bt_iter_pos *_bt_python_create_iter_pos(void);
+struct bt_ctf_iter *_bt_python_ctf_iter_create_intersect(
+	struct bt_context *ctx,
```
jgalar: Indent with 2 tabs.

bindings/python/python-complements.c
```diff
@@ -400,3 +402,22 @@ struct bt_iter_pos *_bt_python_create_iter_pos(void)
 {
 	return g_new0(struct bt_iter_pos, 1);
 }
+
+struct bt_ctf_iter *_bt_python_ctf_iter_create_intersect(
+	struct bt_context *ctx,
```
jgalar: Indent with 2 tabs.

bindings/python/python-complements.h
```diff
@@ -100,3 +100,8 @@ int _bt_python_ctf_clock_set_uuid_index(struct bt_ctf_clock *clock,
 
 /* iterator */
 struct bt_iter_pos *_bt_python_create_iter_pos(void);
+struct bt_ctf_iter *_bt_python_ctf_iter_create_intersect(
+	struct bt_context *ctx,
```
jgalar: Indent with 2 tabs.

bindings/python/reader.py
abusque: Agreed, but I can't just `return None`, given that this is a generator, no?

jgalar: You can just return.

jgalar:
Merged in master. Let's apply the changes (such as intersection_mode as property) on top.

PR #40, opened 2016-04-27
Add properties to TraceCollection for stream intersection

abusque:
This changeset changes the `intersect_mode` attribute of `TraceCollection` to a read-only property, and adds another property `has_intersection` to allow to tell whether a trace has an intersection between its streams via the bindings.


jgalar:
Merged, thanks!

PR #41, opened 2016-04-27
Docs: update debuginfo doc

abusque:
Update the output format in the examples and document the new
--debug-info-target-prefix and --debug-info-full-path command line
options.

Signed-off-by: Antoine Busque <abusque@efficios.com>


jgalar:
Merged, thanks!

PR #43, opened 2016-04-28
Fixes to the test scripts generation for OOT build

mjeanson:
Use AC_CONFIG_FILES to generate all test scripts and let it manage dist and clean which handles in and out of tree builds.


jgalar:
Merged, thanks!

PR #44, opened 2016-04-29
Fix bin info tests

abusque:
This series of commits fixes two possible situations that would arise in bin-info tests when some `debuginfo-data` files were missing: either segfaulting during testing, or having certain tests pass due to falling back to ELF lookup even though we're testing the DWARF lookup.


jgalar:
Merged, thanks!

PR #45, opened 2016-05-05
Fix: swapped libelf and libdw names in configure error message

abusque:
Signed-off-by: Antoine Busque <abusque@efficios.com>


jgalar:
Merged, thanks!

PR #47, opened 2016-05-19
Standardise spelling of debug info

abusque:
This fixes a typo in a configure message when `libelf` or `libdw` were either missing or too old, which referred to the configure argument as `--disable-debuginfo` instead of the correct `--disable-debug-info`. It also standardises the spelling of debug info throughout the code base to be two separate words. Refer to the commit message for details and rationale.


jgalar:
Merged, thanks!

PR #48, opened 2016-10-28
values.h: document that you can call bt_get()/bt_put() on bt_value_null

eepp:


eepp:

PR #49, opened 2016-10-31
tests/lib/Makefile.am: remove -static for debug info tests

eepp:
This flag is not needed for those tests. It breaks the build when
configuring the project with `--disable-static`.


mjeanson:
Addressed for babeltrace2 in https://patchwork.lttng.org/patch/352796/

PR #51, opened 2016-11-01
configure.ac: use $enableval where needed

eepp:


eepp:

PR #53, opened 2016-12-15
Change lttng-live plugin to use externally defined callbacks

Mi-La:
Introduce lttng_live_process_ctf_event called for each CTF event.
Change lttng_live_should_quit to be application defined.


jgalar:
Closing since it is now possible to consumer live events using the 2.x APIs

PR #55, opened 2017-05-24
Fix unknown variable name in Python writer

MMartin5:
Changed variable name 'value' to 'value_field' in order to fix the unknown name error in Python.


jgalar:
I merged an adapted version of this patch in the stable-1.4 and stable-1.5 branches.
The fix is included as part of the Babeltrace 1.4.3 and 1.5.3 releases.

@eepp is currently reworking the bindings for the master branch so I think that code is rewritten for Babeltrace 2.0.

Thanks and sorry for the delay!

PR #56, opened 2017-06-12
Mingw port, second batch

mjeanson:


lib/plugin/plugin.c
```diff
@@ -380,31 +384,72 @@ struct bt_component_class *bt_plugin_find_component_class(
...
+	}
+
+	switch (flag) {
+	case FTW_F:
+		if (name[0] == '.') {
+			/* Skip hidden files */
+			BT_LOGV("Skipping hidden file: path=\"%s\"", file);
+			goto end;
+		}
+		struct bt_plugin_set *plugins_from_file =
```
jgalar: Move the variable declaration at the beginning of this `case` and enclose it within a scope to limit the variable's visibility:

```
[...]
case FOO:
{
    int bar;
}
[...]
```

lib/plugin/plugin.c
```diff
@@ -380,31 +384,72 @@ struct bt_component_class *bt_plugin_find_component_class(
...
+	pthread_mutex_t lock;
+	struct bt_plugin_set *plugin_set;
+	bt_bool recurse;
+} append_all_from_dir_info = {
+	.lock = PTHREAD_MUTEX_INITIALIZER
+};
+
 static
-struct dirent *alloc_dirent(const char *path)
+int nftw_append_all_from_dir(const char *file, const struct stat *sb, int flag,
```
jgalar: Not really a deal breaker, but I am wondering if the `fts` (`fts_open()`, `fts_read()`, etc.) family of functions are available on Windows?

It would be safer to use as currently, a plug-in that would use `nftw()` during its initialization would be problematic as `nftw()` is not guaranteed to be re-entrant.

cli/babeltrace-log.c
```diff
@@ -414,23 +414,22 @@ int main(int argc, char **argv)
...
 		goto error_rmdir;
 	}
-	dir_fd = bt_dirfd(dir);
-	if (dir_fd < 0) {
-		perror("dirfd");
-		goto error_closedir;
-	}
 
-	fd = openat(dir_fd, "datastream", O_RDWR|O_CREAT,
+	file_path = g_build_filename(s_outputname, "datastream", NULL);
```
jgalar: Check the return value for `NULL`.

lib/ctf-writer/writer.c
```diff
@@ -161,22 +164,20 @@ struct bt_ctf_writer *bt_ctf_writer_create(const char *path)
...
-		perror("open");
-		goto error_destroy;
-	}
-
-	writer->metadata_fd = openat(writer->trace_dir_fd, "metadata",
+	writer->metadata_fd = open(metadata_path,
 		O_WRONLY | O_CREAT | O_TRUNC,
 		S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
 
+	// jgalar: Should we handle open() failure here?
```
jgalar: Yes. It is indeed weird that the original doesn't.

lib/ctf-writer/writer.c
```diff
@@ -161,22 +164,20 @@ struct bt_ctf_writer *bt_ctf_writer_create(const char *path)
...
+
+	g_free(metadata_path);
 	return writer;
 
 error_destroy:
-	unlinkat(writer->trace_dir_fd, "metadata", 0);
+	unlink(metadata_path);
```
jgalar: Should only be done after the `open` done on the metadata file.

In fact, that line is not needed as the creation of the metadata file is the last action taken. In other words, if an error occurs, we won't have created it anyway.

lib/ctf-ir/stream.c
```diff
@@ -692,21 +693,23 @@ int create_stream_file(struct bt_ctf_writer *writer,
 	}
 
 	g_string_append_printf(filename, "_%" PRId64, stream->id);
-	fd = openat(writer->trace_dir_fd, filename->str,
+	file_path = g_build_filename(writer->path->str, filename->str, NULL);
```
jgalar: Check return value and
```
ret = -1;
goto end;
```

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+#include <winsock2.h>
+
+#define BT_INVALID_SOCKET INVALID_SOCKET
+#define BT_SOCKET_ERROR SOCKET_ERROR
+#define BT_SOCKET SOCKET
+
+#define MAXNAMLEN 256
+
+static inline
+int bt_socketlib_init(void)
```
jgalar: Can you change the namespace to be common to all socket functions, e.g. `bt_socket_*()`

`bt_socket_init()`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifdef __MINGW32__
+
+#include <winsock2.h>
+
+#define BT_INVALID_SOCKET INVALID_SOCKET
```
jgalar: `BT_SOCKET_INVALID`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+		WSACleanup();
+		ret = -1;
+	}
+
+end:
+	return ret;
+}
+
+static inline
+int bt_socketlib_clean(void)
```
jgalar: `bt_socket_fini()`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+{
+	return WSACleanup();
+}
+
+static inline
+int bt_closesocket(int fd)
```
jgalar: `bt_socket_close(...)`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+{
+	return closesocket(fd);
+}
+
+static inline
+bool bt_sock_interrupted(void)
```
jgalar: `bt_socket_interrupted()`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+{
+	/* There is no equivalent to EINTR in winsock 2.2 */
+	return false;
+}
+
+static inline
+const char *bt_sock_errormsg(void)
```
jgalar: `bt_socket_errormsg()`

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+		errstr = "Winsock library is not ready";
+		break;
+	case WSANOTINITIALISED:
+		errstr = "Winsock library not initialised";
+		break;
+	case WSAVERNOTSUPPORTED:
+		errstr = "Winsock version not supported";
+		break;
+
+	/* getXbyY() errors (already handled in herrmsg):
```
jgalar: format multiline comments like so:
```
/*
 * I LOVE WinSock errors.
 */
```

include/babeltrace/compat/socket-internal.h
```diff
@@ -0,0 +1,395 @@
...
+# ifdef SO_NOSIGPIPE
+#   define MSG_NOSIGNAL SO_NOSIGPIPE
+# elif defined(__MINGW32__)
+#   define MSG_NOSIGNAL 0
+# endif
+#endif
+
+#if defined(MSG_NOSIGNAL)
+static inline
+ssize_t bt_send_nosigpipe(int fd, const void *buffer, size_t size)
```
jgalar: `bt_socket_send_nosigpipe()`

plugins/ctf/lttng-live/viewer-connection.c
```diff
@@ -55,16 +52,16 @@ static ssize_t lttng_live_recv(struct bt_live_viewer_connection *viewer_connecti
 	size_t copied = 0, to_copy = len;
 	struct lttng_live_component *lttng_live =
 			viewer_connection->lttng_live;
-	int fd = viewer_connection->control_sock;
+	BT_SOCKET sock = viewer_connection->control_sock;
 
 	do {
-		ret = recv(fd, buf + copied, to_copy, 0);
+		ret = recv(sock, buf + copied, to_copy, 0);
```
jgalar: Ideally, `recv()` and `send()` would also be wrapped to make the socket library consistent. Not a priority at the moment, though.

compat/compat_mman.c
```diff
@@ -31,36 +31,70 @@
...
 
-static int mmap_infos_cur = 0;
-static int mmap_infos_max = 0;
-static struct mmap_info *mmap_infos = NULL;
+static
+struct mmap_mapping *mapping_create(void)
+{
+	struct mmap_mapping *mapping;
+
+	mapping = malloc(sizeof(struct mmap_mapping));
```
jgalar: Check malloc return value.

compat/compat_mman.c
```diff
@@ -31,36 +31,70 @@
...
+	mapping->map_handle = NULL;
+
+	return mapping;
+}
 
-#define NEW_MMAP_STRUCT_CNT 10
+static
+void mapping_clean(struct mmap_mapping *mapping)
+{
+	if (mapping) {
```
jgalar: Replace by
```
if (!mapping) {
    return;
}

File: compat/compat_mman.c [hunk 18]
```diff
@@ -109,174 +144,128 @@ DWORD map_prot_flags(int prot, DWORD *dwDesiredAccess)
 	return 0;
 }
 
-void *mmap(void *start, size_t length, int prot, int flags, int fd,
+void *mmap(void *addr, size_t length, int prot, int flags, int fd,
```
jgalar: Namespace this compatibility wrapper to `bt_mmap()`.

The function should also be `BT_HIDDEN`.

lib/Makefile.am
```diff
@@ -22,3 +22,18 @@ if BUILT_IN_PYTHON_PLUGIN_SUPPORT
...
 
+# Build a libbabeltrace-ctf library for backwards compatibility.
+# CTF writer used to be in libbabeltrace-ctf in Babeltrace 1, so this
+# file must still exist. As of Babeltrace 2, CTF writer is implemented
+# in libbabeltrace.
+libbabeltrace_ctf_la_SOURCES = babeltrace.c values.c ref.c logging.c
+libbabeltrace_ctf_la_LDFLAGS = $(LT_NO_UNDEFINED) \
+			-version-info $(BABELTRACE_LIBRARY_VERSION)
+
+libbabeltrace_ctf_la_LIBADD = \
```
jgalar: I think we never shipped a CTF writer dll, so we could just forego this lib entirely on Windows.

lib/plugin/plugin.c
mjeanson: The variable?

include/babeltrace/compat/socket-internal.h
mjeanson: The intention was not to namespace this compat code under bt_socket_ since we don't do it for any other compat wrappers. I guess I just suck at naming things, here is a proposal for a more structured naming scheme.

# Use socketlib for library initialisation, which is specific to windows
bt_socketlib_init()
bt_socketlib_fini()

# Use bt_socket_ for custom compat functions
bt_socket_interrupted()
bt_socket_errormsg()

# Use bt_ORIGFUNC_SOMETHING for wrapped functions
bt_close_socket()
bt_send_nosigpipe()

I would keep BT_SOCKET_ERROR and BT_INVALID_SOCKET because the first denotes a socket error return code and the second an invalide socket object id and not an invalide return code.

Thoughts?

lib/Makefile.am
mjeanson: This is not strictly related to the Windows port, as discussed with @eepp the previous behavior of symlinking the library would be quite problematic with distro packages. It seems a lot cleaner to properly build this compat lib.

include/babeltrace/compat/socket-internal.h
jgalar: Agreed for `BT_INVALID_SOCKET` not being an error code, that's a good point.

As for the naming scheme, I would just unify everything under an internal `bt_socket` library to group all of these utilities. In the future, I would move the rest of the socket calls (recv() and others) there too to keep everything consistent.

The rationale behind this is that our current compatibility scheme of having _loose_ `bt_*`-prefixed functions makes it hard to figure where a compatibility wrapper must be used and where something can be used "as-is".

In the end, I typically merge a patch, the build fails on some platform/config, and then I have to figure whether or not a wrapper already exists.

Since sockets seem to require a fair amount of platform-specific code, I think keeping an internal socket library will make everything simpler in the long run.

configure.ac
```diff
@@ -151,6 +149,10 @@ AC_CHECK_FUNCS([ \
 	uname \
 ])
 
+# AC_FUNC_MALLOC causes problems when cross-compiling.
```
mjeanson: AC_FUNC_MALLOC/REALLOC are based on a run-time test which fails when
cross-compiling and we don't use the results from these tests anyway.

I keep them commented because autoscan will recommend we add them.

cli/babeltrace.c
```diff
@@ -620,6 +620,7 @@ int load_dynamic_plugins(struct bt_value *plugin_paths)
 	return ret;
 }
 
+#ifdef BT_BUILT_IN_PLUGINS
```
mjeanson: No problem per say, I thought the log message about loading static plugins when they were not built a bit confusing.

lib/ctf-ir/stream.c
mjeanson: k

include/babeltrace/compat/socket-internal.h
mjeanson: k

compat/compat_mman.c
mjeanson: k

configure.ac
mjeanson: Add comment

jgalar:
Merged, thanks!

PR #62, opened 2017-06-20
Fixes of the day: June 19th

eepp:
Adding what's left to add to log the library (notification objects), and fixing other logging stuff.

Also adding logging to `flt.utils.muxer` which should prove useful for the various errors that can occur there (clock class is not the expected one, and why is that, for example).

Fixing stuff in the AST-to-IR visitor and `notif-iter.c`.


eepp:
Closing because superseded by #63.

PR #64, opened 2017-06-23
Build plugins as unversioned modules

mjeanson:
This will build the plugins as straight .so file without libtools
'lib' prefix and version symlinks.

See this link for more details :

  https://autotools.io/libtool/plugins.html

Signed-off-by: Michael Jeanson <mjeanson@efficios.com>


jgalar:
Merged, thanks!

PR #65, opened 2017-07-04
Fixes of the day: July 4th

eepp:
Changes:

* Logging added to various modules, fixed in other cases.
* Various minor fixes in the library.
* Simplify some sink component classes.
* Add `sink.utils.counter` component class.
* Add discarded events and packets notifications.
* Automatically generate discarded discarded events and packets notifications (by a notification iterator).
* `src.ctf.fs`: name the created stream objects.
* `sink.text.pretty`: print discarded events and packets warning messages.


jgalar:
Merged, thanks!

PR #76, opened 2017-08-04
CTF writer: Drop validation when setting time

tahini:
Fixes bug 1129

When lazily writing events from different streams, the clock
monotonicity is not guaranteed. Checking it globally as it was causes
the faulty event not to be written, though it is fine for its stream.

Signed-off-by: Geneviève Bastien <gbastien@versatic.net>


tahini:
It's probably a drastic solution, but probably easier than chaging the API to set the clock per stream? Anyway, I'm leaving this here and let you decide on its fate.

jgalar:
Closing, see answer at
https://bugs.lttng.org/issues/1129

The check could also be relaxed to enforce the monotonicity on a per-stream basis, but that won't solve the problem for older versions.

PR #80, opened 2017-08-23
Build Python bindings with distutils for consistent installs

mjeanson:
This patch changes the build system used to compile and install the
Python Bindings. Distutils is used to find the right install directory.
When the install directory generated from the install prefix is not in
the Python search path (PYTHONPATH), we print a warning explaining what
can be done to include it.
It uses Distutils which is part of the Python standard library.

Signed-off-by: Michael Jeanson <mjeanson@efficios.com>


jgalar:
Why do we need to move the files to `bindings/python/bt2/bt2`?

mjeanson:
With distutils the python modules files (*.py) needs to be in a directory relative to the setup.py file. The swig files could technically be somewhere else but the swig integration is a bit dumb and it outputs in the directory were the source files are located which means it has to be in the same directory for distutils to pickup the generated .py files.

mjeanson:
The fix commit could be squashed with the first one, I kept it separate for the review.

jgalar:
Merged, thanks!

PR #91, opened 2017-11-22
Fix python bindings errors

pzins:
Fix wrong size for uint64_ft in StreamClass constructor
and wrong class name in the clock offset.


jgalar:
Closing as those two problems were addressed by the following commits:

[Fix: writer.py: make `uint64_ft` a 64-bit (instead of 32-bit) int FT ](https://github.com/efficios/babeltrace/commit/cb3373fec88c4c63645163bcc32f5af3c922c8b9)
[Fix: typo using bt2.ClockClassOffset constructor ](https://github.com/efficios/babeltrace/commit/da2edb5bdef1e06be21c85776381477754710ffe)

Thanks!

PR #93, opened 2017-11-26
Python: Implementation of __repr__ instead of __str__ for Values and Fields

jgalar:
These two commits are only related in that they both address the implementation of the `__repr__` methods in both the Value and Field APIs of the `bt2` python package.

The first commit adds an "Unset" return to all __repr__ implementations of Fields when they are not `set`. This seemed okay after discussion with @eepp, but I'm wondering if there is a canonical representation of an unset object recommended by Python...

We don't want to use `None` as CTF 2 will most likely introduce a `Null` object for which we reserve the semantics of returning `None`.

In the case of Value, I applied the same modifications as were done in Field last week.

These modifications still need tests.

### DIFF
--- a/bindings/python/bt2/bt2/fields.py
+++ b/bindings/python/bt2/bt2/fields.py
@@ -111,6 +111,9 @@ class _NumericField(_Field):
         return float(self._value)
 
     def __repr__(self):
+        if not self.is_set:
+            return 'Unset'
+
         return repr(self._value)
 
     def __lt__(self, other):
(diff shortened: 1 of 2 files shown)


bindings/python/bt2/bt2/fields.py
```diff
@@ -111,6 +111,9 @@ def __float__(self):
         return float(self._value)
 
     def __repr__(self):
+        if not self.is_set:
```
eepp: Since _is set_ is a property of `_Field`, why not use a template method here:

    class _Field(object._Object, metaclass=abc.ABCMeta):
        # ...

        def _repr(self):
            raise NotImplementedError

        def __repr__(self):
            return self._repr() if self.is_set else 'Unset'

        # ...

and replace each specific field class's `__repr__()` with `_repr()`, and remove the redundant `if not self.is_set` check.

jgalar:
Merged.

PR #95, opened 2018-01-11
Fix live-comm: merge TCP socket write-write sequence in a single write

PSRCode:
The live protocol implementation is often sending content
on TCP sockets in two separate writes. One to send a command header,
and the second one sending the command's payload. This was presumably
done under the assumption that it would not result in two separate
TCP packets being sent on the network (or that it would not matter).

Delayed ACK-induced delays were observed [1] on the second write of the
"write header, write payload" sequence and result in problematic
latency build-ups for live clients connected to moderately/highly
active sessions.

Fundamentally, this problem arises due to the combination of Nagle's
algorithm and the delayed ACK mechanism which make write-write-read
sequences on TCP sockets problematic as near-constant latency is
expected when clients can keep-up with the event production rate.

In such a write-write-read sequence, the second write is held up until
the first write is acknowledged (TCP ACK). The solution implemented
by this patch bundles the writes into a single one [2].

[1] https://github.com/tbricks/wireshark-lttng-plugin
Basic Wireshark dissector for lttng-live by Anto Smyk from Itiviti
[2] https://lists.freebsd.org/pipermail/freebsd-net/2006-January/009527.html

Reported-by: Anton Smyk <anton.smyk@itiviti.com>
Signed-off-by: Jonathan Rajotte <jonathan.rajotte-julien@efficios.com>

### DIFF
--- a/plugins/ctf/lttng-live/viewer-connection.c
+++ b/plugins/ctf/lttng-live/viewer-connection.c
@@ -157,6 +157,8 @@ static int lttng_live_handshake(struct bt_live_viewer_connection *viewer_connect
 {
 	struct lttng_viewer_cmd cmd;
 	struct lttng_viewer_connect connect;
+	const size_t cmd_buf_len = sizeof(cmd) + sizeof(connect);
+	char cmd_buf[cmd_buf_len];
 	int ret;
 	ssize_t ret_len;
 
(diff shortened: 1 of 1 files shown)


plugins/ctf/lttng-live/viewer-connection.c
```diff
@@ -1092,27 +1104,32 @@ enum bt_lttng_live_iterator_status lttng_live_get_next_index(struct lttng_live_c
 	struct bt_live_viewer_connection *viewer_connection =
 			lttng_live->viewer_connection;
 	struct lttng_live_trace *trace = stream->trace;
+	const size_t cmd_buf_len = sizeof(cmd) + sizeof(rq);
+	char cmd_buf[cmd_buf_len];
+
```
jgalar: Removed this empty line.

plugins/ctf/lttng-live/viewer-connection.c
```diff
@@ -1348,6 +1368,9 @@ enum bt_lttng_live_iterator_status lttng_live_get_new_streams(
 	struct bt_live_viewer_connection *viewer_connection =
 			lttng_live->viewer_connection;
 	uint32_t streams_count;
+	const size_t cmd_buf_len = sizeof(cmd) + sizeof(rq);
+	char cmd_buf[cmd_buf_len];
+
```
jgalar: Removed this empty line.

jgalar:
Merged in master and stable-2.0. Thanks!

PR #97, opened 2018-03-02
Fix: Python value field on enum type should be str

frdeso:
The current implementation returns the integer value of the numeration
but Babeltrace 1.X was returning the text label.

To ensure compatibility with Babeltrace 1.X, when getting the value of
an enum field, return a string containing the corresponding label for
this value.

This commit also adds a test for this behaviour.


eepp:
Doesn't seem relevant anymore... 😉

PR #98, opened 2018-03-16
Fix: to enable access to sequence_text field type data from API.

romendmsft:
Currently reading from _sequence_text_ field type is not exposed by the API (at least I couldn't find any possible way of doing this).
_sequence_definition->string_ is properly allocated and populated when sequence_text is read from trace events.
However, _bt_ctf_get_string_ does not consider _SEQUENCE_TYPE_ and does not return the string in such type of field.

### DIFF
--- a/formats/ctf/events.c
+++ b/formats/ctf/events.c
@@ -600,12 +600,34 @@ char *bt_ctf_get_string(const struct bt_definition *field)
 {
 	char *ret = NULL;
 
-	if (field && bt_ctf_field_type(bt_ctf_get_decl_from_def(field)) == CTF_TYPE_STRING)
+	if(!field)
+		goto error;
+
+	struct bt_declaration *field_decl = bt_ctf_get_decl_from_def(field);
+	enum ctf_string_encoding sequence_encoding;
+
+	switch(bt_ctf_field_type(field_decl))
+	{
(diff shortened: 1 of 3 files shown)


jgalar:
Hi,

The 1.x release series is in maintenance mode at this point. Is there a reason you can't use Babeltrace 2's API (currently master)?

Thanks.

romendmsft:
Hi,

At the time we took dependency on the Babeltrace API I think version 2 wasn't out from what I remember.
To be honest I didn't have time to go through babeltrace 2 API yet. Is the 2 API drastically changed from 1 API, or is it basically the same? Is there any documentation on the 2 API you could point me too?

Thanks,
Ronan

jgalar:
It is not out yet, but it is in pre-release phase (meaning the final release is coming soon).
The APIs are completely different and we are working on updating the docs.

romendmsft:
I see, I will try taking a look at the code in master when I have some time to see if we can change our code to the API v2. It would be great if you could point me to at least one example API so I can get started more easily.  Thanks.

milianw:
~1.5y later and babeltrace is still not out yet the issue fixed by this patch still remains. Please merge this and release a new version of the stable API - it's packaged by many distros!

eepp:
You should be okay now.

PR #100, opened 2018-06-06
Fix: zero out URL parsing array before use and ensure proper error re…

frdeso:
…porting


frdeso:
I updated the commit message trying to explain what was the problem.

jgalar:
Merged in stable-1.5 with a minor fix: we typically use `{ 0 }` rather than `{0}`.

Thanks!

PR #101, opened 2018-12-10
Fix: do not resolve debug info on build id mismatch

frdeso:
Issue
=====
Babeltrace is showing erroneous debugging information when a trace is
read on a system that has a different version of the binary that of the
binary recording in trace.

This is due to the fact that Babeltrace currently does not confirm that
the binary(or debuglink) before resolving the debugging information.

The build ids in the trace and in the local binary must be the same to
ensure that instruction pointer addresses can be decoded.

Solution
========
Now when setting the build id, confirm that the trace build id and the
binary build id match. Once the match is confirmed, set the
`file_build_id_matches` field in the corresponding the `bin_info`
structure. This way when decoding the events, we can quickly check if
the debugging information is available for that binary.

Drawback
========
None.

### DIFF
--- a/include/babeltrace/bin-info.h
+++ b/include/babeltrace/bin-info.h
@@ -62,6 +62,8 @@ struct bin_info {
 	int dwarf_fd;
 	/* Denotes whether the executable is position independent code. */
 	bool is_pic:1;
+	/* Denotes whether the build id in the trace matches to one on disk. */
+	bool file_build_id_matches:1;
 	/*
 	 * Denotes whether the executable only has ELF symbols and no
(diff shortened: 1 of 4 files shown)


frdeso:
This PR is for the stable-1.5 branch.

eepp:
This bug report is no longer relevant.

Babeltrace 2 now verifies that the build ID recorded in the trace matches the build ID of the binary found on the filesystem before resolving debug information. If they don't match, we log a message and the lookup fails gracefully instead of showing erroneous debug info.

Thank you for your interest in the Babeltrace project!

PR #103, opened 2019-03-12
Fix: set effective byte order of FT copy _and_ original

eepp:


dschaefer:
Cool, I tried this out on my test and it does fix my problem. Thanks!

eepp:
This bug is effectively mitigated because:

* Field types default to `BT_CTF_BYTE_ORDER_NATIVE`.
* Serialization always falls back to the native byte order of the trace when a field type has `BT_CTF_BYTE_ORDER_NATIVE`.
* Even if the user keeps original field type references and creates fields from them, serialization will use the correct byte order from the trace

Therefore this bug report is no longer relevant in practice, though the implementation differs from what was described in the patch notes.

Closing.

PR #104, opened 2019-03-29
debug-info: Fix: not printing src field in for loops

frdeso:
Issue
=====
When debugging information is available, Babeltrace does not print the
`src` field of a tracepoint properly when tracepoint is within a for
loop.

  #include "tp.h"
  int main(int argc, char *argv[])
  {
      tracepoint(my_app, empty);
      for(int i = 0; i < 1; i++) {
          tracepoint(my_app, empty);
      }
      return 0;
  }

The problem derives from the DWARF information generated by the
compiler. The compiler may place the DWARF content of the for loop in
`DW_TAG_lexical_block` which the current debug-info design doesn't
expect.

Here is a summarized version of DWARF tree containing the first
tracepoint in the code above:
  DW_TAG_subprogram
    DW_TAG_inlined_subroutine
      /* tracepoint callsite info */

Here is a summarized version of the DWARF tree containing the tracepoint
in the for loop:
  DW_TAG_subprogram
    DW_TAG_lexical_block
      DW_TAG_inlined_subroutine
        /* tracepoint callsite info */

The current implementation doesn't expect the presence of a
`DW_TAG_lexical_block` entry as child of the `DW_TAG_inlined_subroutine`
entry and won't look any further down that branch to find the source
information for the current tracepoint. This results in the Babeltrace
not finding the line number for such callsite and thus leaving the `src`
field empty.

Solution
========
When iterating over the Debugging Information Entries, if the current
entry contains the address to resolve _but_ is not a of the type
`DW_TAG_inlined_subroutine` try to iterate over the children of that
entry.

Known drawbacks
===============
None.

### DIFF
--- a/include/babeltrace/dwarf.h
+++ b/include/babeltrace/dwarf.h
@@ -120,6 +120,15 @@ BT_HIDDEN
 void bt_dwarf_die_destroy(struct bt_dwarf_die *die);
 
 /**
+ * Indicates if the debug information entry `die` has children DIEs.
+ *
+ * @param die	bt_dwarf_die instance
+ * @returns	0 if the die no child, 1 otherwise
+ */
+BT_HIDDEN
+int bt_dwarf_die_has_children(struct bt_dwarf_die *die);
+
+/**
  * Advance the debug information entry `die` to its first child, if
  * any.
  *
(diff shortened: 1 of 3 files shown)


jgalar:
Merged, see e63652425414abf7a1ed670f042d244a31019dc5.

PR #106, opened 2019-05-17
export bt_ctf_field_type_{integer_get_base,string_get_encoding}

milianw:
These two functions are useful when you want to write a custom CTF analyzer using libbacktrace

### DIFF
--- a/formats/ctf/events.c
+++ b/formats/ctf/events.c
@@ -600,12 +600,34 @@ char *bt_ctf_get_string(const struct bt_definition *field)
 {
 	char *ret = NULL;
 
-	if (field && bt_ctf_field_type(bt_ctf_get_decl_from_def(field)) == CTF_TYPE_STRING)
+	if(!field)
+		goto error;
+
+	struct bt_declaration *field_decl = bt_ctf_get_decl_from_def(field);
+	enum ctf_string_encoding sequence_encoding;
+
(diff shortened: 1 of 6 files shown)


milianw:
I just realize there's `bt_ctf_get_encoding` and `bt_ctf_get_int_base` which can be used as alternatives for now. I still think it would be good to have these functions available though.

jgalar:
Hi Milian,

The functions you exported are exclusively used by the CTF-Writer API.
For the reader-side of the Babeltrace 1.x API, see `include/babeltrace/ctf/events.h`.

Both `bt_ctf_get_encoding` and `bt_ctf_get_int_base` are available in the "reader" header. The 1.x is not fresh to my mind, but I thing that to determine if an array or a sequence contains a string, you need to:
1. access the first element of the array/sequence
2. get the element's declaration (see `bt_ctf_get_decl_from_def()`)
3. use that declaration with `bt_ctf_get_encoding` and `bt_ctf_get_int_base` to emulate the logic of `bt_int_is_char()` (see `types/integer.c`)

Does that work?

jgalar:
Note that there is no way to access the `alignment` property of a `bt_declaration`. Hence, that part of the check is not possible without adding a suitable accessor.

eepp:
Babeltrace 1.x isn't relevant anymore.

PR #108, opened 2019-05-31
bugfix babeltrace live was failing to connect if hostname had _

stephening:
The reason was that a sscanf was failing the pattern match. After adding _ to the set, I was able to successfully view a live LTTng trace stream on a system that had _ in the hostname.


eepp:
Not relevant anymore in Babeltrace 2.

PR #110, opened 2020-01-15
`swig -swiglib` returns multiple paths

jamores:
When trying to build python-bindings under windows, using MSYS2, mutliple paths were returned when populating SWIG_LIB by `swig -swiglib`, breaking MAKEFILE generation in the process as SWIG_LIB was written incorrectly:

> SWIG_LIB=<first_path>
<second_path>

Adding `tail -1`, just the last path is preserved and SWIG_LIB is written properly @ generated MAKEFILE.
Not sure if this is the best solution as I'm no autotools expert.

### DIFF
--- a/m4/ax_pkg_swig.m4
+++ b/m4/ax_pkg_swig.m4
@@ -121,7 +121,7 @@ AC_DEFUN([AX_PKG_SWIG],[
                                 m4_ifval([$3],[$3],[])
                         else
                                 AC_MSG_CHECKING([for SWIG library])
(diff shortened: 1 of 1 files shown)


simark:
Hi @jamores,

The file you modified is maintained in the [autoconf archive](https://github.com/autoconf-archive/autoconf-archive), and we merely copy it in our repository.  Please send a patch to them, they will help you find a proper solution for your problem.  Once a fix is merged in the autoconf archive repo, then we can sync our copy of the file with them.

PR #111, opened 2020-02-11
ctf: Support enum values as bit fields

tahini:
When enum mapping values are all powers of 2, the value is considered as
a bit field value and the corresponding mappings should be all those
with this bit set to 1.

Example babeltrace output of such a field (block_rq* kernel events):

[13:15:49.024354958] (+0.000003868) wilbrod block_rq_complete: { cpu_id = 4 },
    { dev = 8388624, sector = 375490176, nr_sector = 360, error = 0,
      rwbs = ( "RWBS_FLAG_READ", "RWBS_FLAG_RAHEAD" : container = 12 ) }

Signed-off-by: Geneviève Bastien <gbastien@versatic.net>

### DIFF
--- a/src/lib/trace-ir/field-class.c
+++ b/src/lib/trace-ir/field-class.c
@@ -391,6 +391,8 @@ struct bt_field_class *create_enumeration_field_class(
 		goto error;
 	}
 
+	enum_fc->is_bit_field_enum = true;
+
 	BT_LIB_LOGD("Created enumeration field class object: %!+F", enum_fc);
 	goto end;
 
@@ -558,8 +560,8 @@ bt_field_class_enumeration_unsigned_get_mapping_labels_for_value(
 				BT_INTEGER_RANGE_SET_RANGE_AT_INDEX(
 					mapping->range_set, j);
 
-			if (value >= range->lower.u &&
-					value <= range->upper.u) {
+			if ((enum_fc->is_bit_field_enum && ((range->lower.u & value) != 0)) ||
+				(value >= range->lower.u &&	value <= range->upper.u)) {
 				g_ptr_array_add(enum_fc->label_buf,
 					mapping->label->str);
 				break;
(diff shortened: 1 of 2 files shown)


src/lib/trace-ir/field-class.c
```diff
@@ -669,6 +671,23 @@ add_mapping_to_enumeration_field_class(struct bt_field_class *fc,
 	BT_LIB_LOGD("Added mapping to enumeration field class: "
 		"%![fc-]+F, label=\"%s\"", fc, label);
 
+	// See if this remains a bit field enum field
+	if (enum_fc->is_bit_field_enum) {
+		if (range_set->ranges->len != 1) {
```
compudj: What is this doing ? Does it exclude enums which have duplicated labels for the same bit ? If it is the case, is it really what we want here ?

src/lib/trace-ir/field-class.c
```diff
@@ -669,6 +671,23 @@ add_mapping_to_enumeration_field_class(struct bt_field_class *fc,
...
+			enum_fc->is_bit_field_enum = false;
+		} else {
+			const struct bt_integer_range *range = (const void *)
+				BT_INTEGER_RANGE_SET_RANGE_AT_INDEX(
+					range_set, 0);
+			if ((range->lower.i != range->upper.i) ||
```
compudj: This excludes enums that contain at least one range which contains more than a single value. (perhaps a bit of comments ?)

src/lib/trace-ir/field-class.c
```diff
@@ -669,6 +671,23 @@ add_mapping_to_enumeration_field_class(struct bt_field_class *fc,
...
+				!(range->lower.i > 0 && (range->lower.i & (range->lower.i - 1)) == 0)) {
```
compudj: Do we really want to use lower.i here ? And why exclude values with the 64-th bit set (which appear negative) ? I suspect we want to deal with lower.u and upper.u instead, and don't care about sign.

tahini:
After discussion on irc, this PR will move to gerrit and...

1- This feature should go in the 'pretty' sink of babeltrace instead of the main library

2- A single label can have multiple values, as long as they are powers of 2, it's acceptable. A single value can be used for more than one label. These cases need to be covered.

PR #113, opened 2020-09-14
configure.ac: add --{dis,en}able-tests option

ffontaine:
Add an option to allow the user to disable tests. Indeed, tests are not
always needed (especially when cross-compiling for an embedded target)
and consume time and CPU ressources on builders

To keep current behavior, tests are enabled by default

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -404,6 +404,12 @@ AC_ARG_ENABLE([man-pages],
   [enable_man_pages=yes]
 )
 
+# Enabled by default
+AC_ARG_ENABLE([tests],
+  [AS_HELP_STRING([--disable-tests], [Do not build and install tests])],
+  [], dnl AC_ARG_ENABLE will fill enable_tests with the user choice
+  [enable_tests=yes]
+)
 
 # Set automake variables for optionnal feature conditionnals in Makefile.am
 AM_CONDITIONAL([ENABLE_PYTHON_BINDINGS], [test "x$enable_python_bindings" = xyes])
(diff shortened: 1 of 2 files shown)


compudj:
The build failure should be fixed, and I don't think disabling tests because they do not build is a good argument for not building tests.

About the cross-compiling argument, indeed the tests would not be run on the host platform, but those would be useful on the target, so building them is still relevant.

ffontaine:
Indeed, you're right that the test should be fixed, for example by renaming `struct user`.

However, I still feel that disabling tests can be useful for the end user as we're wasting time and CPU ressources to build them. There was a already the possibility for the user to disable documentation so I was thinking that it would be acceptable to also have an option to disable tests. However, that's your project so it's up to you to decide what is good. I can close this PR and send an other one to fix the build failure or perhaps you want to fix it yourself? In this case, I'll retrieve the commit to fix the build failure on our autobuilders.

compudj:
Having a configure option for disabling build of tests makes sense overall, and the reasons you outline in your reply justify making this configurable. If you change your commit message to justify it based on time and cpu resources of builders, I think it would really improve it.

If you have time to submit a fix renaming struct user it would be very welcome and appreciated.

eepp:
Can we move this to Gerrit?

Personally I'm not against a configuration option to disable building tests. It's a part of the project which one could want to build or not, just like we have those options for Python bindings, C API documentation, Python bindings documentation, manual pages, etc.

Especially since testing might require additional dependencies (not sure it's Babeltrace 2's case; I think we just don't _run_ some tests if Python is not available, for example).

Distribution packaging is a prime example of wanting to build the project without tests, and for some distributions (embedded), without documentation.

ffontaine:
I updated the PR as requested. I'm not familiar with your contribution process. Do I need to do something to move it on Gerrit?

eepp:
> Do I need to do something to move it on Gerrit?

I don't think we have a clear guide explaining the process yet, so [I did it for you](https://review.lttng.org/c/babeltrace/+/4075).

eepp:
@ffontaine You can sign into our Gerrit through GitHub and take part in the review process.

ffontaine:
Thanks a lot

ffontaine:
I'm closing this PR as I saw that it was rejected on gerrit

PR #117, opened 2021-05-26
ctf-writer: Fix list of reserved keywords

berolinux:
The missing comma between "void" and "_Bool" causes the compiler to
treat it as one string ("void_Bool"), missing both "void" and "_Bool"
as real reserved keywords.

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>


jgalar:
Merged manually in master and stable-2.0. Thanks!

PR #119, opened 2022-05-08
Fix the build with slibtool

orbea:
gentoo bug: https://bugs.gentoo.org/777444

Note:
* This PR is based against the `stable-1.5` branch.
* The current `master` branch is not affected.
* Gentoo still ships Babletrace 1 and does not yet have Babletrace 2. I have no knowledge when this will change.

The Babeltrace 1 build fails when using slibtool (https://dev.midipix.org/cross/slibtool) instead of GNU libtool with many multiple definition warnings. This appears to be a case of overlinking and can be solved by cleaning up the `.la` dependencies for the build. Slibtool is more sensitive to user error while GNU libtool will silently hide some of the worst bugs which is how this goes unnoticed. With this patch it now works with both slibtool and GNU libtool on my system. I am not sure if you wish to carry this patch too, but I am sharing it in case you find it valuable.

### DIFF
--- a/formats/ctf-text/Makefile.am
+++ b/formats/ctf-text/Makefile.am
@@ -12,9 +12,7 @@ libbabeltrace_ctf_text_la_LDFLAGS = \
 	$(LD_NO_AS_NEEDED) -version-info $(BABELTRACE_LIBRARY_VERSION) \
 	types/libctf-text-types.la
 
-libbabeltrace_ctf_text_la_LIBADD = \
-	$(top_builddir)/lib/libbabeltrace.la \
-	$(top_builddir)/formats/ctf/libbabeltrace-ctf.la
+libbabeltrace_ctf_text_la_LIBADD =
 
 if ENABLE_DEBUG_INFO
 libbabeltrace_ctf_text_la_LIBADD += $(top_builddir)/lib/libdebug-info.la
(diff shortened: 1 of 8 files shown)


sskras:
(from a passer-by)

Well, slibtool got updated in the meantime. Maybe it's worth retesting?

eepp:
> Well, slibtool got updated in the meantime. Maybe it's worth retesting?

Please validate this and reopen if there's still an issue.

That being said, Babeltrace 1.x has been EOL for many years now, therefore we won't put much effort into the matter.

sskras:
@eepp, are the latest 1.x changes here, please: https://github.com/efficios/babeltrace/commits/stable-1.5 ?

sskras:
@orbea, on 2022-05-09 22:21:15 +0000 you wrote:

> dev-util/babeltrace: Fix overlinking
>     
> When building with slibtool the build fails with many multiple
> definition errors. This is because of overlinking internal .la
> files where GNU libtool somehow obscures the issue.
>     
> Note the newer Babeltrace 2 which is not yet in Gentoo does
> not seem to have any slibtool related build issues.

Does Gentoo still builds **Babeltrace 1** ?

eepp:
> Does Gentoo still builds **Babeltrace 1** ?

Seems like [yes](https://packages.gentoo.org/packages/dev-util/babeltrace)?

orbea:
Gentoo now has 2.1.2 as well, I don't personally use babeltrace and only looked at this because a Gentoo issue was made.

PR #120, opened 2025-03-21
Make bt_field_blob_get_length return size_t instead of uint64_t

kraj:
Fixes errors e.g.
| ../../git/src/cpp-common/bt2/field.hpp:1139:82: error: non-constant-expression cannot be narrowed from type 'std::uint64_t' (aka 'unsigned long long') to 'size_type' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]
|  1139 |         return {internal::CommonBlobFieldSpec<LibObjT>::data(this->libObjPtr()), this->length()};
|       |                                                                                  ^~~~~~~~~~~~~~
| ../../git/src/plugins/ctf/common/src/msg-iter.cpp:744:56: note: in instantiation of member function 'bt2::CommonBlobField<bt_field>::data' requested here
|   744 |     std::memcpy(&this->_stackTopCurSubField().asBlob().data()[_mCurBlobFieldDataOffset],
|       |                                                        ^
| ../../git/src/cpp-common/bt2/field.hpp:1139:82: note: insert an explicit cast to silence this issue
|  1139 |         return {internal::CommonBlobFieldSpec<LibObjT>::data(this->libObjPtr()), this->length()};
|       |                                                                                  ^~~~~~~~~~~~~~
|       |                                                                                  static_cast<size_type>( )

### DIFF
--- a/src/cpp-common/bt2/field.hpp
+++ b/src/cpp-common/bt2/field.hpp
@@ -1136,7 +1136,7 @@ public:
 
     bt2s::span<typename internal::CommonBlobFieldSpec<LibObjT>::Data> data() const noexcept
     {
-        return {internal::CommonBlobFieldSpec<LibObjT>::data(this->libObjPtr()), this->length()};
+        return {internal::CommonBlobFieldSpec<LibObjT>::data(this->libObjPtr()), static_cast<std::size_t>(this->length())};
     }
 
     std::uint64_t length() const noexcept


eepp:
Thank you.

If you don't mind, could you [create an issue on our official bug tracker](https://bugs.lttng.org/issues/new) instead? If you can't, I will.

May I also get your `configure` command line?

This specific fix isn't the right one. `bt2::CommonBlobField::length()` has every right to return `std::uint64_t`: it wraps a `uint64_t`-returning function. The constructor of `bt2s::span` expects `std::size_t`; that's where you want to cast `this->length()`.

kraj:
@eepp this issue happens when building for `riscv32` targets using clang-20 compiler. I think it will also fail on other 32bit platforms too.

kraj:
I do not have account on the bug tracker and creating new is sent to overlords for approval :)

eepp:
In fact don't worry about the bug tracker, this is so simple we'll fix it now.

The patch is fine. The "right" cast is probably something like

    typename bt2s::span<typename internal::CommonBlobFieldSpec<LibObjT>::Data>::size_type

(lol) but whatever.

However can you move this to a [Gerrit](https://review.lttng.org/) change? You already have an account there.

kraj:
done here - https://review.lttng.org/c/babeltrace/+/14316

PR #121, opened 2025-08-21
src: explicitly only build shared plugins

rossburton:
For some reason, libtool builds static libraries for libs which have been marked as modules. These are pointless: you can't dlopen() a static library.

Avoid wasting compile time and disk space by passing -shared to libtool so that the static libraries are not built.


eepp:
I missed this for some reason.

Would you please submit to our [Gerrit Code Review system](https://review.lttng.org/)? We'll review once there.

eepp:
Closing this because there's no activity.

Feel free to submit as a Gerrit change!

PR #122, opened 2025-11-27
Fix: trimmer: directly use provided value if integer

fmoessbauer:
The trimmer plugin allows to specify the filter begin and end as signed integer as well as timestring. When providing the value as integer, we cannot simply convert the number to a string, as the date parser does not accept this format. This is anyways not needed, as the API clearly states that signed integers (typed) are directly used as time in nanoseconds.

We fix this by directly processing the value in case it is a signed integer.

Fixes: 7de0e49a7 ("Adapt `flt.utils.trimmer` to current API")


kienanstewart:
Review link: https://review.lttng.org/c/babeltrace/+/15723

eepp:
Reviewed on Gerrit now: closing.

PR #123, opened 2025-11-28
sink.ctf.fs: add support to create LTTng index

fmoessbauer:
The LTTng index provides data to efficiently seek in an CTF file by denoting the sizes and offsets of the packets. This helps when working with large traces that are only partially read (like in trace-compass). The source.ctf.fs component is already capable of reading this index, however the corresponding sink component lacks support for writing it.

We add support to generate the index file, if explicitly requested by the user.


eepp:
See https://lists.lttng.org/pipermail/lttng-dev/2025-December/031032.html.

